Issue #2, opened 2019-11-16
Add possibility to expose multiple identities under the same domain name


jloic:
Since virtually every other did-resolver is able to expose multiple identities, I'd like to use a single domain to expose multiple DID. Accepting a path after the domain name would go a long way, but if it clashes then we can split the path with something different.
Example:
 "did:web:my.domain.com|{uuid}" that would resolve to 
 - https://my.domain.com/did/{uuid}

gribneau:
I've suggested adding a parameter to the well-known URL rather than changing it.
https://github.com/w3c-ccg/did-method-web/pull/4

dmitrizagidulin:
I think the DID:web method should support _both_:
* Discovery of a DID document given a bare domain name (with no path) (through either the `.well-known/` or the DNS TXT mechanisms).
* Ability for any arbitrary path to house a DID document. I want to be able to use `did:web:example.com/dmitri/did.json` as my DID.

gribneau:
I agree, particularly with the arbitrary path case above.

I would be happy to contribute a definition for that, if you like. (I've actually been working on it separately.)

dmitrizagidulin:
@gribneau Go for it, that would be great! :)

dmitrizagidulin:
(The discussion continues on PR #5.)

OR13:
Demo: https://did-web.web.app/ (uses 302 redirect for both variants).

dmitrizagidulin:
Implemented in PR #5, closing.

PR #4, opened 2019-11-29
Support multiple DIDs per domain with a parameter.

gribneau:
If we extend the specification very slightly to support a parameter on the well-known URL, a site with many users could resolved DIDs for each user.

This is in response to issue #2. The solution differs slightly from the request to maintain the well-known url.


gribneau:
This PR has been superseded by https://github.com/w3c-ccg/did-method-web/pull/5 at the rquest of 
@dmitrizagidulin .

PR #5, opened 2019-12-03
Add support for optional full path in the method specific identifier.

gribneau:
This simple extension retains support for well-known URLs on bare domains and adds the option to specify a full path.

It implies that Dmitri's DID will be added to w3c-ccg.github.io.

```
did:web:w3c-ccg.github.io/dmitri/did.json
```

```
https://w3c-ccg.github.io/dmitri/did.json
```

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -83,21 +83,29 @@ DID, after the prefix, is specified below.
 ## Method Specific Identifier
 
 The method specific identifier is a fully qualified domain name that is secured
-by a TLS/SSL certificate. The formal rules describing valid domain name syntax
-are described in [RFC 1035](https://tools.ietf.org/html/rfc1035), [RFC
-1123](https://tools.ietf.org/html/rfc1123), and [RFC
-2181](https://tools.ietf.org/html/rfc2181).
+by a TLS/SSL certificate with an optional path to the DID document. The
+formal rules describing valid domain name syntax are described in
+[RFC 1035](https://tools.ietf.org/html/rfc1035),
+[RFC 1123](https://tools.ietf.org/html/rfc1123), and
+[RFC 2181](https://tools.ietf.org/html/rfc2181).
 
 The method specific identifier must match the common name used in the SSL/TLS
-certificate, and it must not include IP addresses, port numbers, or directories
-behind the top-level domain extension.
+certificate, and it must not include IP addresses or port numbers. Directories
+and subdirectories may optionally be included, delimited by colons rather
+than slashes.
 
 	web-did = "did:web:" domain-name
 
-### Example
+    web-did = "did:web:" domain-name * (":" path)
+
+### Examples
 
 ```
 did:web:w3c-ccg.github.io
+```
+
+```
+did:web:w3c-ccg.github.io:user:alice
 ```
 
 ## JSON-LD Context Definition
@@ -128,24 +136,42 @@ Creating a DID is done by
   service
 1. creating the DID document JSON-LD file including a suitable keypair, e.g.,
   using the Koblitz Curve, and storing the `did.json` file under the well-known
-  URL.
+  URL to represent the entire domain, or under the specified path if many DIDs
+  will be resolved in this domain.
 
 For the domain name `w3c-ccg.github.io`, the `did.json` will be available under
 the following URL:
 
 ```
+did:web:w3c-ccg.github.io
+```
+
+```
 https://w3c-ccg.github.io/.well-known/did.json
+```
+
+If an optional path is specified rather the bare domain, the `did.json`
+will be available under the specified path:
+
+```
+did:web:w3c-ccg.github.io:user:alice
+```
+
+```
+https://w3c-ccg.github.io/user/alice/did.json
 ```
 
 ### Read (Resolve)
 
 The following steps must be executed to resolve the DID document from a Web DID:
 
-1. Parse the fully qualified domain name from the identifier.
+1. Replace ":" with "/" in the method specific identifier to obtain the fully
+  qualified domain name and optional path.
 2. Generate an HTTPS URL to the expected location of the DID document by
-  prepending `https://` and appending `/.well-known/did.json` to the domain
-  name.
-3. Perform an HTTP `GET` request to the URL using an agent that can
+  prepending `https://`.
+3. If no path has been specified in the URL, append `/.well-known`.
+4. Append `/did.json` to complete the URL.
+5. Perform an HTTP `GET` request to the URL using an agent that can
   successfully negotiate a secure HTTPS connection, which enforces the security
   requirements as described in
   [Security Considerations](Security-Considerations).
@@ -203,6 +229,30 @@ It is recommended to adhere to
 latest recommendations for hardening TLS configurations.
 
 Delete action can be performed by domain name registrars or DNS lookup services.
+
+### Optional Path Considerations
+
+When optional paths to DID documents are used to resolve documents rather than
+bare domains, verification with signed data proves that the entity in control of
+the file indicated in the path has the private keys. It does not prove that the
+domain operator has the private keys.
+
+This example:
+
+```
+did:web:example.com:u:bob
+```
+
+resolves to the DID document at:
+
+```
+https://example.com/u/bob/did.json
+```
+
+In this scenario, it is probable that example.com has given user Bob control
+over the DID in question, and proofs of control refer to Bob rather than all
+of example.com.
+
 
 ## Reference Implementations
 
--- a/index.html
+++ b/index.html
@@ -1432,7 +1432,7 @@ pre .property::before, pre .property::after {
   <div class="head">
    <p data-fill-with="logo"></p>
    <h1 class="p-name no-ref" id="title">did:web Decentralized Identifier Method Specification</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2020-03-18">18 March 2020</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2020-03-19">19 March 2020</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
--- a/index.html
+++ b/index.html
@@ -1167,6 +1167,18 @@ Possible extra rowspan handling
 		margin-left: auto;
 		margin-right: auto;
 	}
+	.overlarge {
+		/* Magic to create good table positioning:
+		   "content column" is 50ems wide at max; less on smaller screens.
+		   Extra space (after ToC + content) is empty on the right.
+
+		   1. When table < content column, centers table in column.
+		   2. When content < table < available, left-aligns.
+		   3. When table > available, fills available + scroll bar.
+		*/ 
+		display: grid;
+		grid-template-columns: minmax(0, 50em);
+	}
 	.overlarge > table {
 		/* limit preferred width of table */
 		max-width: 50em;
@@ -1176,7 +1188,6 @@ Possible extra rowspan handling
 
 	@media (min-width: 55em) {
 		.overlarge {
-			margin-left: calc(13px + 26.5rem - 50vw);
 			margin-right: calc(13px + 26.5rem - 50vw);
 			max-width: none;
 		}
@@ -1184,14 +1195,12 @@ Possible extra rowspan handling
 	@media screen and (min-width: 78em) {
 		body:not(.toc-inline) .overlarge {
 			/* 30.5em body padding 50em content area */
-			margin-left: calc(40em - 50vw) !important;
 			margin-right: calc(40em - 50vw) !important;
 		}
 	}
 	@media screen and (min-width: 90em) {
 		body:not(.toc-inline) .overlarge {
 			/* 4em html margin 30.5em body padding 50em content area */
-			margin-left: 0 !important;
 			margin-right: calc(84.5em - 100vw) !important;
 		}
 	}
@@ -1212,7 +1221,7 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version e86d5b1b47a216fa66165c234892adc259212183" name="generator">
+  <meta content="Bikeshed version 576ef4f5126e3fb83bf07d9bbe1a64f99aa4944e" name="generator">
   <link href="https://github.com/w3c-ccg/did-method-web" rel="canonical">
 <style>/* style-md-lists */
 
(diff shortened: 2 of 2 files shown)


dmitrizagidulin (approved):
👍 from me, thanks @gribneau!  @awoie - does this addition work for you?

awoie:
I see one issues with having full paths and we have to at least point that out under security considerations what could be dangerous about it. Resolving a did:web and verifying some piece of signed data would no longer proof that the did:web controller is the controller of the domain. @dmitrizagidulin in your example, you are the controller but proofing control of that did won't proof that you under control of github.io.

awoie:
@dmitrizagidulin uPort is implementing an approach where they provide an indication in their UI that a selective disclosure request came from a particular domain, e.g., github.io.

awoie:
@gribneau can you please provide the corresponding language in the security considerations section?

awoie (changes requested):
I see one issue with having full paths and we have to at least point that out under security considerations what could be dangerous about it. Resolving a did:web and verifying some piece of signed data would no longer proof that the did:web controller is the controller of the domain. In your example, you are the controller but proofing control of that did won't proof that you are under control of github.io.

dmitrizagidulin:
@awoie good points, that stuff is important to consider.

So, I think that having paths does not necessarily increase security risk, in the sense of, it has very similar semantics as URLs do in general. For example:

Resolving a web DID without a path: `did:web:example.com`:
- 2 ways to discover the DID, a) through `/.well-known/`, and b) through DNS TXT record, right?
- The security implication is that the did controller (given that they authenticate, with the appropriate key in the doc) controls the domain. (Or at very least, controls the `/.well-known/` top level dir OR the DNS records.)

Resolving a web _with_ a path: `did:web:example.com/users/dmitri/did` 
- The security implication is just that the did owner has control of _just that directory_, so it could be like a multi-tenant user profile thing.

I think developers are used to reasoning about the two kind of URLs? One that represents the organization that controls the domain, and the other just a given user on that domain?

gribneau:
I've added a section in Security Considerations to indicate that full paths very probably indicate user control of keys rather than domain operator control of keys.

Does this meet the need @awoie ?

awoie:
@dmitrizagidulin According to the DID spec, "/" is not in the range of `method-specific-id`

```
did                = "did:" method-name ":" method-specific-id
method-name        = 1*method-char
method-char        = %x61-7A / DIGIT
method-specific-id = *idchar *( ":" *idchar )
idchar             = ALPHA / DIGIT / "." / "-" / "_"
```

So, `did:web:example.com/users/dmitri/did` needs to be URL encoded like `did:web:example.com%2Fusers%2Fdmitri%2Fdid`.

awoie (changes requested):
Can you pls update the PR to address the URL encoding part: https://github.com/w3c-ccg/did-method-web/pull/5#issuecomment-564024166

dmitrizagidulin:
@awoie 
> According to the DID spec, "/" is not in the range of `method-specific-id`

Right right - so, in this case, the / is part of the DID Path actually, not the id.

gribneau:
Percent encoding is pretty ugly.

```
method-specific-id = *idchar *( ":" *idchar )
```

We could use `:`, translate to `/`, and require percent encoding in the odd case that someone uses a `:` in a url.

How does that sound @awoie ?

dmitrizagidulin:
@gribneau - wait, I don't think we need to encode at all. We're using it as a DID url path (like any url path)

gribneau:
That's right - I recall looking this up.

https://www.w3.org/TR/did-core/#path

Example 7:

```
did:example:123456/path
```

We're using:

```
did:web:example.com/path/to/did.json
```

awoie:
So, the DID will be `did:web:example.com`in both cases? I think that does
not work for the following reason...

Example 1:
`did:web:github.com/scammerorganization/scammingdids/blob/master/did.json`

Example 2:
`did:web:github.com`

That means that both refer to the same DID but they are clearly not under
the control of the same entity. Would you agree with that?

Oliver

awoie:
@gribneau What problem do you want to solve with using the path notation? Regardless of the domain binding, I am asking because I am concerned that having the path version and the .well-known version of did:web will have two separate security guarantees:
- `did:web:example.com` -> it is very likely that the did doc can only be updated by the owner of the domain, i.e., one entity
- `did:web:example.com/my/oliver/did.json`-> the owner of example.com granted me access to that directory. We add another middle man (in addition to domain registrars) that could manipulate my did doc anytime.

They might even fall into two separate W3C DID method rubrics.

gribneau:
> What problem do you want to solve with using the path notation?

This thread began with a member of the public noting that most DID methods support resolution of many DIDs rather than just one.

I agree with that observation, and am trying to make it possible for a website to offer DID resolution for an arbitrarily large number of users.

> * `did:web:example.com` -> it is very likely that the did doc can only be updated by the owner of the domain, i.e., one entity
> * `did:web:example.com/my/oliver/did.json`-> the owner of example.com granted me access to that directory. We add another middle man (in addition to domain registrars) that could manipulate my did doc anytime.

The first example resolves to the DID document of the domain itself. The second example resolves to a DID document hosted on that domain, probably of a user.

I do not think any person with a technical background would confuse these identifiers:

`did:web:example.com`

`did:web:example.com/alice/did.json`

If the concern is that the DID Path is used inappropriately, we can make a very simple modification to resolve this.

`did:web:example.com/alice/did.json`

becomes

`did:web:example.com:/alice/did.json`

This would then be more consistent with the use of "should" in the DID Path definition, and requires a very minor adjustment.

> They might even fall into two separate W3C DID method rubrics.

If the `web` method is intended to resolve precisely one DID document per domain, rather than supporting resolution of people's DID documents via the web generally (as it surely appears based on the nomenclature selected), I can create an alternative method.

Please advise @dmitrizagidulin @awoie .

dmitrizagidulin:
@awoie 
> So, the DID will be `did:web:example.com`in both cases? I think that does
not work for the following reason...
> Example 1: `did:web:github.com/scammerorganization/scammingdids/blob/master/did.json`
> Example 2: `did:web:github.com`
> That means that both refer to the same DID but they are clearly not under the control of the same entity. Would you agree with that?

Ok, so, we've got several things to unpack here. 

One is - just like `example.com/user/one` and `example.com/` are not the same URL, they would not result in the same DID. The [_authority_ part of the URL](https://tools.ietf.org/html/rfc3986#section-3.2) is the same, in each case, true. (Both in the `did:web` and regular url case). But the URL (and DID) is different.

Two - how do we deal with different URLs with the same _authority_ part of the URL potentially representing different controllers / different trust levels? I would argue that this sort of calculation is built into the URL already. All developers and most users are familiar with the trust level difference of `github.com/` and `github.com/someorg/`. 
What's even more interesting, is - the type of sites that allow users to publish arbitrary content (like Github, and Wordpress), explicitly vary the authority depending on the user. Meaning, WordPress specifically puts user IDs into the authority part of the URL, to differentiate `alice.wordpress.com` and `wordpress.com`. Note that Github does this as well! You _can't_ access a DID file just with a raw github URL, `github.com/scammer-org/did`. You can only access it via Github.io, which is very careful with the authority -- `scammer-org.github.io/did`.

> We add another middle man (in addition to domain registrars) that could manipulate my did doc anytime.

Three - right, ok, so thirdly we have a concern about integrity protection of the contents of the DID Documents (can the web host mess with the contents whenever they want?). This is a legit concern, regardless of whether we allow paths to be a part of the `did:web` method or not. For this, we have to employ other mechanisms -- we can require (or strongly recommend) that the _DID Document itself_ be signed by the controller. (Preventing tampering). And we can point to the Hashlink mechanism, so that people can refer to a particular version of the DID Document without any concern of tampering by the host.

gribneau:
The three part breakdown from @dmitrizagidulin above raises issues that should be addressed, but I think it is worth noting that use cases for web resolution of many user DIDs are emerging in mainstream conversations, and this specification is very close to solving the basic user identity problem in a portable fashion.

Jack Dorsey announced yesterday that Twitter will fund development of open standards for distributed social media. This will require verifiable identity, and we can easily imagine platforms at the scale of Twitter generating DIDs per user to sign content that will be distributed arbitrarily.

For example, Jack's DID Identifier could be:

`DID:WEB:twitter.com/jack/did.json`

For a scheme like this to be practical, the specification would need to remain flexible enough to support either user control for platforms that place control of the DID document in the hands of the users, or platform control for use cases in which the platform creates and manages the DID documents but wishes to identify each user discretely.

I have no reason to believe that Twitter would necessarily go this route, but the use case presents itself clearly in Jack's announcement:

https://twitter.com/jack/status/1204766078468911106

dmitrizagidulin:
Ok, so, we have two main obstacles here. One is technical/ABNF related, and one is conceptual (what sort of trust model are we trying to convey).

#### 1) Technical Obstacle - ABNF
So, like @awoie pointed out, the current DID Core data model / ABNF rules for the DID uri does not allow the `id` attribute to have path (or query or fragment) components.
So, for example, currently this would not work:

```
did:web:example.com/users/dmitri/did.json

-> resolves to DID Document:

{
  "id": "did:web:example.com/users/dmitri/did.json",
  "authentication": [ ... ],
  // ...
```

^ that wouldn't work, because the Subject (identified by the `id` property) isn't supposed to have a path component.
(Although I brought this up as an issue in the DID WG - see [my comment on #183](https://github.com/w3c/did-core/issues/183#issuecomment-593703715).)

**Potential solutions:**

**1.a)**: Campaign to change the DID Document data model / ABNC rules :)

**1.b)**: Allow the `did:web` method to encode the path component of the https url in the _authority_ segment of the DID.

So, an https URL of `https://example.com/dmitri/did.json` would be encoded as:

`did:web:example.com%2Fdmitri%2Fdid.json`.

(Although, as @gribneau points out, percent-encoding is kind of clunky, UI wise, but we could potentially think of some other encoding.)

#### 2) Strategic/Policy Question - SHOULD we allow this at all?
Even more important than the previous item is of course the question "Is this a good idea at all?" (because it's a safe bet that if yes, we can solve obstacle 1)).

Part of the reason (I suspect) that @awoie and others are objecting to this PR, is that currently, we have a tidy trust model:

**One domain (the _hostname_ part of the URL) === One and Only One DID**

So, the current implication is - a `did:web` url has the full support/authority of the domain behind it.

Which means that the current answer to "how do we provide DIDs for our users, on our web site" is "give them subdomains". (Especially since now, wildcard SSL certs are cheap, and Lets Encrypt allows wildcart certs for free.)

So here, the main question is - do we _want_ to support multiple DIDs on a hostname (via some solution to item 1 above)?
And if yes, what's the justification - is it that giving users subdomains is prohibitive for some reason?

dmitrizagidulin:
My current stance on that question 2) is - I'm not sure. I _can_ see arguments for allowing multiple DIDs on the same hostname -- just due to IT Department difficulties with provisioning new SSL certs for subdomains.
For example, say a company wants to give DIDs to all of their staff, at `did:web:bigcompany.com/staff/12345.json` and so on. It might be a deal-breaker for some potential users to be able to provision a subdomain for each staff member.

Let's wait to hear from various implementers?

gribneau:
I agree that we can find a solution to issue 1 if there is agreement on issue 2.

> #### 2) Strategic/Policy Question - SHOULD we allow this at all?
> Even more important than the previous item is of course the question "Is this a good idea at all?" (because it's a safe bet that if yes, we can solve obstacle 1)).
> 
> Part of the reason (I suspect) that @awoie and others are objecting to this PR, is that currently, we have a tidy trust model:
> 
> **One domain (the _hostname_ part of the URL) === One and Only One DID**

We do indeed have a tidy model linking a DID to a domain. The name `DID:WEB`, though, lends itself to a more expansive interpretation, as most sites have multiple users, and "identity" is more readily associated with persons that sites.

> So here, the main question is - do we _want_ to support multiple DIDs on a hostname (via some solution to item 1 above)?
> And if yes, what's the justification - is it that giving users subdomains is prohibitive for some reason?

It is feasible to support potentially thousands of subdomains with wildcard certificates, but I think it is more likely that `bigcompany.com` would simply disregard this standard and do something simpler.

My position on the issue is that the specification would support many more use cases if we resolve many DIDs in each domain, and that this would enhance the value of the specification.

dmitrizagidulin:
Since my previous comment yesterday, I've grown more convinced that this is a needed feature (to support multiple DIDs per hostname).

The ability to create subdomains is a fairly advanced skill set, _considerably_ more so than just registering a domain name.

OR13:
Seems like its fine to do :`did:web:example.com:/directory/subdirectory/` if you are enforcing the file name is `did.json` it does not need to included in the did... Pretty common in schools to see:

`did:web:university.example.com:/student/~jsmith/`

As for the concern over the trust domain issues, as long as :

`did:web:example.com:/directory/subdirectory/` and `did:web:example.com` are different DIDs I think its fine to have one be domain admin and the other be student... obviously if you control the domain, you control 100% of `did:web`... regardless of which type of did you have... this is similar approach to what we took with Github DID... which you can see people are using by querying github: https://github.com/search?q=ghdid

`https://github-did.com/api/v1/did/did:github:OR13`

We convert the github username to a url to a file in a github repo...

You would be doing the same thing, just for websites other than github.

dmitrizagidulin:
@OR13 
> Seems like its fine to do :did:web:example.com:/directory/subdirectory/ if you are enforcing the file name is did.json it does not need to included in the did

Agreed, yeah. Except we need to find a way to encode the slashes (due to the ABNF issue I mention above). And so far, I'm kind of liking the idea of using `:` (in the did authority segment) to encode slashes. So, specifically:

```
https://example.com/directory/subdirectory/did.json   ->  did:web:example.com:directory:subdirectory (did.json implied)
```

gribneau:
@dmitrizagidulin 

> So, specifically:
> 
> ```
> https://example.com/directory/subdirectory/did.json   ->  did:web:example.com:directory:subdirectory (did.json implied)
> ```

This is quite close to one of my initial drafts. If there is a consensus on this approach, I'll be happy to resolve the conflicts and update the PR.

dmitrizagidulin:
@awoie - Would you agree with this direction?

awoie:
@dmitrizagidulin two things... 

- there seems to be consensus that the domain and path-style DIDs are two distinct DIDs and have different security guarantees (domain vs user). Is this correct? Can we please provide the two examples from @OR13 and provide explicit language that both DIDs are not equivalent. 

- Further, I think the examples still contain "/" in the method-specific-id. I believe there was an issue with that and the solution was to use ":" instead? 

@mirceanis ^^

dmitrizagidulin:
@awoie sounds good!

dmitrizagidulin:
@gribneau - thanks for the updates! (Do you mind rebasing on the latest index.bs, so that there's no conflicts?)

@awoie - 
> there seems to be consensus that the domain and path-style DIDs are two distinct DIDs and have different security guarantees (domain vs user). Is this correct? Can we please provide the two examples from @OR13 and provide explicit language that both DIDs are not equivalent.

Correct. Does the language in the [Optional path considerations](https://github.com/w3c-ccg/did-method-web/pull/5/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR233) section work, in terms of providing explicit language?

dmitrizagidulin:
Thanks again to @gribneau for rebasing, and all your work on this, and thank you to @awoie for excellent discussion. Merging.

mirceanis:
Besides the [uPort/DIF implementation](https://github.com/decentralized-identity/web-did-resolver), what are other implementations that need to be notified of this change in spec?

dmitrizagidulin:
@mirceanis Great question. I'm working on an implementation as well. The only other one I can think of is @peacekeeper's universal resolver?

mirceanis:
@dmitrizagidulin, the uniresolver uses the uport driver which uses the implementation that is now hosted in DIF.

Issue #6, opened 2019-12-09
Add language that explicitly deters IDPs (Facebook, Google, MSFT etc.) from abusing subdomain-based IDs


csuwildcat:
I feel we should add language to the spec that specifically warns against and deters use of did:web with subdomains in ways that are not in keeping with the purpose if DIDs. For example: a large IDP may try to create subdomains on its centralized IDP service domain and 'issue' DIDs to users/customers, which would essentially lock their DID to that provider, the exact opposite of what DIDs were designed to do.

gribneau:
I am not sure I fully understand the concern.

Let's examine a use case involving an email service idpmail.com with users Alice and Bob. If I am reading the above correctly, it would be abusive for this IDP to resolve:

`did:web:alice.idpmail.com` and
`did:web:bob.idpmail.com`

to enable Alice and Bob to expose public keys so that they could send, receive, verify or decrypt signed or encrypted email.

Similarly, user Charlie with a private server could participate by exposing:
`did:web:charlie.privatemail.com`

Why would this be abusive?

OR13:
I recommend closing this, I don't see this an issue, in fact it would be awesome to see more adoption by larger tech companies of did:web... for example, github already supports something very similar....

https://github.com/OR13.gpg

wouldn't it be better if they did this instead?

did:web:github:OR13 :)

gribneau:
I agree with closing the issue, and I agree that it would be great if github embraced DID:WEB, which now does what they need.

:)

OR13:
I am currently squatting on did:github, would love to hand that over to them as well...  https://github-did.com/resolver

PR #9, opened 2020-10-19
Add language about correlation risk to individuals

msporny:


index.bs
```diff
@@ -189,6 +190,36 @@ longer publicly available due to any other means.
...
 ## Security and Privacy Considerations
 
+### High Risk of Pervasive Tracking
+
+The reliance of this DID Method on traditional HTTP-based Web Architecture can
+enable the same sorts of privacy-invasive tracking technologies that are in
+widespread use today (2020). For example, verification of a Verifiable
+Presentation might result in resolving a did:web DID by the verifier. This
+enables the DID Method registry to know when and where every did:web DID on the
+ledger is used, which is useful when tracking behaviour of the DID subject.
```
rhiaro, line 200: What is "DID method registry" in this context? Does that mean the server which hosts the DID doc? I'd change the language to avoid confusion with [this DID method registry](https://www.w3.org/TR/did-spec-registries/#did-methods). We could say "Verifiable Data Registry" to be consistent with the DID Core spec?

In this case I also wouldn't use the word "ledger" here.

(I haven't reviewed the rest of the spec to make sure language aligns with the latest in DID Core, so I'll do that too.)

rhiaro, line 200:
```suggestion
enables the DID Method registry to know when and where every did:web DID on the
ledger is used, which enables anyone with access to the sever to track behaviour of the DID subject.
```

index.bs
```diff
@@ -189,6 +190,36 @@ longer publicly available due to any other means.
...
+
+This concern is not necessarily problematic for public entities, such as
+government organizations or some corporations, but is highly problematic
+for individuals from a privacy perspective.
+
+It is strongly advised that a privacy violation will occur if this DID Method
+is used for DIDs where the DID subject requires privacy. Examples of
```
rhiaro, line 207:
```suggestion
is used for DIDs where the DID subject requires privacy and the Verifiable Data Registry infrastructure (e.g. the domain name and Web server) is not under direct and exclusive control of the DID subject. Examples of
```

index.bs
```diff
@@ -189,6 +190,36 @@ longer publicly available due to any other means.
...
+inappropriate use of this DID Method include, but are not limited to:
+government-issued DIDs to non-employees, corporate-issued DIDs to non-employees,
+and website-issued DIDs to people that are not in control of the website
+infrastructure.
+
+To put a finer point on the privacy concerns of this particular DID Method.
```
rhiaro, line 213:
```suggestion
```

(this doesn't stand on its own as a sentence)

index.bs
```diff
@@ -189,6 +190,36 @@ longer publicly available due to any other means.
...
+Government-issued did:web DIDs to citizens enable pervasive tracking of
+citizens. Corporate-issued did:web DIDs to individuals enable pervasive tracking
+of those individuals. Website-issued did:web DIDs to users enable pervasive
+tracking of those users.
+
+It is strongly advised that governments and corporations strongly advise
+individuals pick a DID Method that is more auditable and privacy-preserving
+than this DID Method.
```
rhiaro, line 221:
```suggestion
It is strongly advised that individuals do not use this DID method unless they have absolute confidence in their control over the underlying infrastructure of the implementation; that is, that the individual controls and has access to the domain name and Web server being used. It is strongly advised that individuals do not agree to or make use of a did:web DID that is issued on their behalf by a centralized authority which can control the underlying infrastructure without notice or auditability.

It is strongly advised that where governments and corporations need to issue DIDs for individuals (as opposed to companies, departments or assets for example) they use a DID method that is more auditable and privacy-preserving than did:web.
individuals pick a DID Method that is more auditable and privacy-preserving
than this DID Method.
```

rhiaro (changes requested):
Thanks! This more or less address https://github.com/w3c-ccg/did-method-web/issues/6 I think.

index.bs
dmitrizagidulin: Minor typo with 'sever'

dmitrizagidulin:
@msporny +1 to this PR overall (we've been meaning to add this section for a while).

We should get more specific with the warnings, though.

> For example, verification of a Verifiable Presentation might result in resolving a did:web DID by the verifier. This enables the ~~DID Method registry~~ Verifiable Data Registry to know when and where every did:web DID on the
ledger is used, which is useful when tracking behaviour of the DID subject.

The "when and where every did:web DID ... is used" is not quite right. We should say something more like:

"For example, the verification of a Verifiable Presentation might result in a verifier resolving a did:web DID. As with any HTTP request, this resolution enables the Verifiable Data Registry (the server where the DID is hosted) to track a number of things, both about the verifier and the DID itself. (Incidentally, these concerns apply to any DID methods accessed through an http-based Universal Resolver of any sort hosted on a public website.)

DID: 
* the server can track the frequency with which a DID is requested (which allows for various statistics gathering regarding relative popularity of DIDs).
* the exact time the verification was requested (which, combined with the next section, can help pinpoint the identity of the verifier).

Verifier:
* the server can track the requesting IP address of the verifier, which frequently enables geolocation-by-IP or even the exact identity of a well-known institution by their IP.
* the server can track any URL query parameters that are added to the DID by the requester (for example, some search engines, when directing a user to a website, pass along the search terms that were used to find it).
* if the verification request was performed using a typical web browser, the server can track even more about the verifier:
    - The `User-Agent` (as well as any other HTTP headers), which often enables browser fingerprinting attacks.
    - Any cookies the browser sends along with the request, which are frequently used for tracking.
"

index.bs
OR13: +1 to using the term VDR.

OR13:
unclear what changes need to be accepted to see this merged.

dmitrizagidulin:
@OR13 - I'd like the warnings to be more specific. (discussing this with @msporny)

OR13:
PR is stale and should be closed.

msporny:
Yes, it's stale. If we close this, the text should be preserved and integrated into the new spec.

OR13:
https://github.com/w3c-ccg/did-method-web/issues/29

This PR should be closed if the requested changes cannot be implemented in a timely manner.

@dmitrizagidulin @awoie I recommend adding a "pending-close" / 1 week tag.

OR13:
@msporny this PR has conflicts and changes requested, if they are not addressed in 1 week it will be closed, if you wish for these concerns to be addressed, please open issues to track them.

OR13:
Closing the PR, feel free to open another one at any time.

PR #21, opened 2020-11-07
Convert from bikeshed to respec (substance unchanged)

rhiaro:


dmitrizagidulin (approved):
👍 
Thank you @rhiaro!

Issue #27, opened 2021-04-20
"publicKey" not defined in did-core


letmaik:
Example 1 uses `"publicKey"` which is not defined in https://www.w3.org/TR/did-core/

Am I missing something?

dmitrizagidulin:
Hi @letmaik -- no, you're absolutely right, we just need to update the did:web spec.
(`publicKey` got renamed to `verificationMethod`, in the DID Core spec.)

dmitrizagidulin:
@letmaik - This should be addressed by PR https://github.com/w3c-ccg/did-method-web/pull/28

PR #28, opened 2021-04-23
Update to reflect DID Core changes (publicKey, context).

dmitrizagidulin:
Addresses issue #27.

Updates example DID Document to match pre-CR changes to the DID Core spec.

* Change DID core context to `https://www.w3.org/ns/did/v1`
* Rename `publicKey` to `verificationMethod` (per did core spec).

### DIFF
--- a/index.html
+++ b/index.html
@@ -109,18 +109,17 @@ Add non-ethereum example keys (ed25519, RSA, etc)
 
         <pre class="example" title="Example did:web DID document">
 {
-  "@context": "https://w3id.org/did/v1",
+  "@context": "https://www.w3.org/ns/did/v1",
   "id": "did:web:example.com",
-  "publicKey": [{
-       "id": "did:web:example.com#owner",
-       "type": "Secp256k1VerificationKey2018",
-       "owner": "did:web:example.com",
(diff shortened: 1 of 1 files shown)


OR13 (approved):

index.html
```diff
@@ -109,18 +109,17 @@ <h2>
...
-  "publicKey": [{
-       "id": "did:web:example.com#owner",
-       "type": "Secp256k1VerificationKey2018",
-       "owner": "did:web:example.com",
-       "ethereumAddress": "0xb9c5714089478a327f09197987f16f9e5d936e8a"
+  "verificationMethod": [{
+     "id": "did:web:example.com#owner",
+     "type": "Secp256k1VerificationKey2018",
+     "owner": "did:web:example.com",
+     "ethereumAddress": "0xb9c5714089478a327f09197987f16f9e5d936e8a"
```
OR13, line 118: please remove this example, and include examples from:

https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/implementations/did-key-2020-db.json#L42

OR13: since the proposed context change makes this invalid json-ld.

dmitrizagidulin: @OR13 ah, right, thanks!

PR #30, opened 2021-05-12
Remove the spec Editors from Digital Bazaar.

msporny:
This PR removes the spec Editors from Digital Bazaar as other Editors would like to step forward to move the specification forward at a faster rate than it's been progressing.


dmitrizagidulin (approved):
:+1:

rhiaro (approved):

kimdhamilton:
😢

PR #32, opened 2021-05-13
mesur.io would like to be an editor for did:web

mprorock:
mesur.io based on our work in this area would like to be an editor for the specification


dmitrizagidulin (approved):
👍

PR #38, opened 2021-08-19
update text to specify how ports should be handled

### LINKED ISSUE #7: How to handle hostnames with ports?
Now that we support paths (as of PR #5) in our `did:web` URLs, _and_ we're using the `:` to encode the `/` characters for paths, this poses another challenge.

Since we're using the `:` character for paths, what do we do about the _actual_ most common intended purpose of that character, which is to specify a port number?

So, specifically, say I am a developer who has just fired up a test server on their local machine, which is running on `https://localhost:8443` (the `https` is deliberate of course - I made a self-signed cert for it and everything). This kind of thing happens all the time (it's happening to me _right now_ :) ).

If there's a `did:web` document residing on that domain (say in `/.well-known/did.json`), what will that URL look like? According to our rules so far:

```
did:web:localhost:8443
```

Except now we're using `:` (in the `did-specific-identifier` portion of the url) to encode path fragments. So that URL would "decode" to `https://localhost/8443`. Not what we want.

So, what are people's thoughts on how to best handle this? @awoie, @OR13 ?

kdenhartog:
fixes #7

### DIFF
--- a/index.html
+++ b/index.html
@@ -197,9 +197,10 @@ The formal rules describing valid domain name syntax are described in
         </p>
         <p>
 The method specific identifier MUST match the common name used in the SSL/TLS
-certificate, and it MUST NOT include IP addresses or port numbers. Directories
-and subdirectories MAY optionally be included, delimited by colons rather
-than slashes.
+certificate, and it MUST NOT include IP addresses. A port MAY be included and
+the colon MUST be percent encoded to prevent a conflict with paths. Directories
+and subdirectories MAY optionally be included, delimited by colons rather than
+slashes.
         </p>
 
         <pre class="nohighlight">
@@ -211,6 +212,8 @@ web-did = "did:web:" domain-name * (":" path)
 did:web:w3c-ccg.github.io
 
 did:web:w3c-ccg.github.io:user:alice
+
+did:web:example.com%3A3000
         </pre>
       </section>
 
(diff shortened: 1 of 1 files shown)


mirceanis:
Looks good, but I think percent encoding should happen regardless of port, to create valid DIDs from hostnames with unicode chars

kdenhartog:
> Looks good, but I think percent encoding should happen regardless of port, to create valid DIDs from hostnames with unicode chars

Do you have an example I should include around this? Are you thinking about percent encoding international Unicode characters?

mirceanis:
I don't have an example, but I guess this spec should also cover internationalized domain names.
Normally they are encoded using punycode, but I suspect that is not as readily available as percent encoding.

gribneau:
> If the domain contains a port percent decode the colon.

If we specify percent encoding as a part of the standard, I think it should be applied generally.

It's been a while, but I do recall some conversation on this topic, and I think views were positive overall.

mirceanis:
My comment is also based on those discussions.
Those discussions also mentioned percent encoding the path components.

kdenhartog:
Ok, I think I understand. I'll update the text to reflect that the entire domain gets percent encoded. I'm less certain we want to do this with paths though because percent encoding can pick up `/` which could potentially lead to weird behaviors. How about we modify this text for domains only for now and we can discuss paths further to come up with a solid solution for how that might work.

mirceanis:
I should have been more clear when i mentioned path components. I meant the sections of the path string that go between the slashes :)
But I agree, this can be done later.

dmitrizagidulin (approved):
:+1:  from me. (I think we should tackle Unicode / internationalized domain names in a separate PR).

kdenhartog:
3 months later and I still haven't finished updating this PR with the suggestions from @mirceanis 

I'm aware of what he means now after our team implemented this further. ~~I just need to update the text to make it more clear. I'll add this to my list of higher priority things to get finished up next week.~~

I'm going to open a second PR to add that additional text instead.

OR13:
When can we merge this? would the folks commenting here, please approve or request changes?

OR13 (approved):

dmitrizagidulin:
Merging this PR, as it's an important detail to have in the spec.
We'll address percent-encoding path fragments in a separate PR.

Issue #42, opened 2021-11-11
Proposal to resolve representation type, file extensions, and mime type issues


gribneau:
We have a number of open issues regarding different representation types, file extensions, and mime types that have been floating around for a while now (#8 #15 #20 #41). I think we are also not in compliance with the [IANA Considerations](https://w3c.github.io/did-core/#iana-considerations) section of the core spec at this point.

Resolving these issues gracefully seems likely to compel us to modify the simple resolution schema. A file named `did.json` should not contain, for example, CBOR or YAML, and setting the appropriate content type without changing the file extension strikes me as a poor workaround.

Beginning with the core spec, we should be minimally supporting at least two representations:

```
.didjson
application/did+json

.didjsonld
application/did+ld+json
```

We can accomplish this by changing the URL resolution logic and relying on webserver configuration to return an appropriate file and set the proper content type by resolving to a directory containing multiple representations rather than a file.

```
did:web:example.com:charlie

https://example.com/charlie/
```

In this scenario, the webserver would be configured to recognize a number of index files, each with the appropriate file extension. These file extensions would also need to map to the appropriate content types.

```
charlie/did.didjson
charlie/did.didjsonld
charlie/did.cbor
```

We could require that simple json be accepted for all resolutions to support dumb webservers, but recognize the `accept` request header to prioritize other formats on servers that are capable of handling that logic.

```
Accept: application/did+ld+json
Accept: application/cbor
Accept: application/did+json
```

The additional complexity should all be entirely optional so that a simple webserver hosting a simple `.didjson` file will work today, but there is a path available to those who need other representations.

This would require modifying the URL resolution logic (and also the existing resolvers that have been implemented), but I think it would make more sense in the end.

What does everyone think?

dmitrizagidulin:
@gribneau - I agree, we do need to address this (at least, to bring it into compliance with the DID Core spec, IANA registration wise).

(I agree with you that we should support both "dumb" webservers that are just hosting files, but also smart ones that can perform content negotiation.)

So, we've got three distinct cases for did resolution.

1. Resolving a DID for the top level domain (like `did:web:example.com`) by appending a well-known `did.*` filename. (Incidentally, we really should reconcile this with (or at very least mention) the [Well Known DID Configuration](https://identity.foundation/.well-known/resources/did-configuration/) spec.)
[...]

PR #43, opened 2021-11-19
Reorganize, streamline, and extend did:web. #42

gribneau:
I've reorganized the document, extended the resolution logic, added server
configuration considerations, and updated RFC references.

There are still comments in this document referencing various issues, but I have
moved them to sections dedicated to discussing the issues raised. We should 
reach a consensus on paths forward for these. I think most of them have been 
resolved by simply recognizing (and calling out) the reality of resolution over
https and can be removed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -197,9 +197,10 @@ The formal rules describing valid domain name syntax are described in
         </p>
         <p>
 The method specific identifier MUST match the common name used in the SSL/TLS
-certificate, and it MUST NOT include IP addresses or port numbers. Directories
-and subdirectories MAY optionally be included, delimited by colons rather
-than slashes.
+certificate, and it MUST NOT include IP addresses. A port MAY be included and
+the colon MUST be percent encoded to prevent a conflict with paths. Directories
+and subdirectories MAY optionally be included, delimited by colons rather than
+slashes.
         </p>
 
         <pre class="nohighlight">
@@ -211,6 +212,8 @@ web-did = "did:web:" domain-name * (":" path)
 did:web:w3c-ccg.github.io
 
 did:web:w3c-ccg.github.io:user:alice
+
+did:web:example.com%3A3000
         </pre>
       </section>
 
@@ -284,6 +287,15 @@ did:web:w3c-ccg.github.io:user:alice
  -> https://w3c-ccg.github.io/user/alice/did.json
           </pre>
 
+          <p>
+If an optional port is specified on the domain, the port colon splitting the
+host and the port MUST be percent encoded to prevent collision with the path.
+          </p>
+
+          <pre class="example nohighlight" title="Creating the DID with optional path and port">
+did:web:example.com%3A3000:user:alice 
+ -> https://example.com:3000/user/alice/did.json
+          </pre>
         </section>
 
         <section>
@@ -299,6 +311,9 @@ DID:
             <li>
 Replace ":" with "/" in the method specific identifier to obtain the fully
 qualified domain name and optional path.
+            </li>
+            <li>
+If the domain contains a port percent decode the colon.
             </li>
             <li>
 Generate an HTTPS URL to the expected location of the DID document by


index.html
```diff
@@ -187,25 +190,28 @@ <h2>
...
         </p>
         <p>
 The method specific identifier MUST match the common name used in the SSL/TLS
 certificate, and it MUST NOT include IP addresses. A port MAY be included and
 the colon MUST be percent encoded to prevent a conflict with paths. Directories
 and subdirectories MAY optionally be included, delimited by colons rather than
-slashes.
+slashes. Colons in directories and subdirectories MUST be percent encoded, if
+present, to avoid being translated into slashes. A filename MAY optionally be
+included as the final section of the colon delimited method specific identifier.
```
dmitrizagidulin:
```suggestion
slashes. A filename MAY optionally be included as the final section of the colon 
delimited method specific identifier. All path fragments (directories and 
filenames) MUST be percent encoded.
```
I think we should emphasize that all characters in path fragments are percent-encoded, not just colons.

index.html
```diff
@@ -309,21 +316,30 @@ <h3>
 
           <ol>
             <li>
-Replace ":" with "/" in the method specific identifier to obtain the fully
-qualified domain name and optional path.
+Replace ":" with "/" in the method specific identifier, exclusive of any fragment,
```
dmitrizagidulin: I'm not sure I understand the 'exclusive of any fragment' part -- what do you mean by that?

index.html
```diff
@@ -332,42 +348,41 @@ <h3>
...
-from modification by an attacker.
-          </p>
-
-          <p class="note" title="HTTP API">
-There is no HTTP API specified for the update process leaving programmatic
-registrations and management to be defined by each implementation.
+          <p>
+In the event that multiple representations of the same DID are present in a
+directory, updates of all representations MUST be made when any representation is
+updated. Key material and service endpoints MUST be consistent across representations
```
dmitrizagidulin: Oooh, great point here.

dmitrizagidulin (approved):
👍 
Great work, @gribneau! Thank you.

index.html
gribneau: I am referring to anything following `#` as the fragment.

In this example:
`did:web:example.com:alice#key9`

I refer to `key9` as the fragment.

We've not made any reference to fragments before now, and perhaps we can exclude that and simplify the language a bit further.

index.html
gribneau: Colons are a special case. If a filename or path contains a literal `:`, that would be translated into `/` if not percent encoded.

If the colon was not our delimiter, it would not need percent encoding.

gribneau:
Within the context of this PR, I think it would be beneficial to address a number of issues that have been noted in the text of the specification with an eye to ensuring that our explicit handling of those issues is adequate to remove them from the text and close them.

These appear on lines 462, 466, 487, and 582.

That informational section is intended to address those issues (and others), and avoid making did:web something it cannot be without a dramatic increase in complexity. This method should support DID delivery by static webservers in the simplest possible fashion.

There is also an ongoing discussion about the language describing potential DNS vulnerabilities and mitigation in  #45, and we could pick up #46 and #44 as well.

Ideally, I'd like this PR to clean up the text and clear out as many outstanding issues as possible.

index.html
dmitrizagidulin: @gribneau 
> If the colon was not our delimiter, it would not need percent encoding.

Ah, we still would, though, because directory and file names can contain characters that are not allowed by the DID spec. So it's not just the colon.

index.html
```diff
@@ -214,35 +220,22 @@ <h2>
...
-        <pre title="Web DID method JSON-LD context">
-{
-    "@context":
-    {
-        "ethereumAddress": "https://github.com/uport-project/ethr-did-resolver#ethereumAddress",
-        "Secp256k1VerificationKey2018": "https://github.com/uport-project/ethr-did-resolver#Secp256k1VerificationKey2018",
-        "Secp256k1SignatureAuthentication2018": "https://github.com/uport-project/ethr-did-resolver#Secp256k1VerificationKey2018",
-    }
-}
+did:web:example.com:christine:111.json
```
OR13: can we not mix these file name based variations into this PR?

Too many changes at once, IMO.

index.html
```diff
@@ -260,42 +253,56 @@ <h3>
...
-For example, for the domain name `w3c-ccg.github.io`, the `did.json` will be
-available under the following URL:
+The well-known URL for a domain name is the simplest example:
           </p>
-
-          <pre class="example nohighlight" title="Creating the DID">
+          <pre class="example nohighlight" title="DID at well known path">
 did:web:w3c-ccg.github.io
- -> https://w3c-ccg.github.io/.well-known/did.json
+ -> https://w3c-ccg.github.io/.well-known/did/
```
OR13: I suggest we not co-mingle these changes with the other cleanups, also this is a breaking change that would make existing implementations non conformant... cc @kdenhartog

index.html
```diff
@@ -376,17 +391,117 @@ <h3>
...
 
+      <section>
+      <h2>
+Webserver Configuration
+      </h2>
+      <h3>
+MIME Types
+      </h3>
+      <p>
+Webservers MUST be configured to associate proper content types with registered
```
OR13: I would prefer to see `application/json` as the only defined format first... and then attempt to address these other representations separately... too many changes in this PR.

OR13 (changes requested):
Let's split out the file system and content type changes from the rest of this PR.

I'm generally in favor of the changes but concerned we are doing too much at once, and especially worried about the interop and breaking changes associated with the proposed representation changes.

index.html
mprorock: +1 we need to break this apart into multiple PRs

mprorock (changes requested):
We should break this apart, at least by grammatical cleanup, followed by any PRs that may introduce breaking changes

index.html
gribneau: The `:` that appears in the domain segment of the method specific identifier in the case that a port is specified must be percent-encoded to avoid being rewritten as `/`.

e.g.
```
did:web:example.com:1443
---> https://example.com/1443/

vs.

did:web:example.com%3A1443
---> https://example.com:1443/.well-known/did/
```

index.html
gribneau: I removed this in #47.

index.html
gribneau: A reduced PR with a clean history appears at #47.

index.html
gribneau: In #47, I reduced to `application/json` and `application/did+json` to preserve the logic of shifting to index files instead of hardcoded `/did.json`.

I also removed references to JSON-LD.

This presents us with a path forward to comfortably support other representations in the future without embracing the complexity today.

gribneau:
There is a clean PR at #47 reflecting feedback above.

PR #50, opened 2021-12-14
Simplify did resolution, drop .well-known/ mechanism.

dmitrizagidulin:
Based on the discussion in PR https://github.com/w3c-ccg/did-method-web/pull/47, and the consensus to simplify the DID-to-web-url translation rules by dropping the `.well-known/` mechanism, and appending `/did.json` directly.

Previously:

```
did:web:example.com  --->   https://example.com/.well-known/did.json
```

Now (as of this PR):

```
did:web:example.com  --->   https://example.com/did.json
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -261,8 +261,8 @@ service
             </li>
             <li>
 creating the DID document JSON-LD file including a suitable keypair, e.g.,
-using the Koblitz Curve, and storing the <code>did.json</code> file under the
-well-known URL to represent the entire domain, or under the specified path if
+using the Koblitz Curve, and storing the <code>did.json</code> file in the root
+directory to represent the entire domain, or under the specified path if
 many DIDs will be resolved in this domain.
             </li>
           </ol>
@@ -274,7 +274,7 @@ available under the following URL:
 
           <pre class="example nohighlight" title="Creating the DID">
 did:web:w3c-ccg.github.io
- -> https://w3c-ccg.github.io/.well-known/did.json
+ -> https://w3c-ccg.github.io/did.json
           </pre>
 
           <p>
@@ -293,7 +293,7 @@ host and the port MUST be percent encoded to prevent collision with the path.
           </p>
 
           <pre class="example nohighlight" title="Creating the DID with optional path and port">
-did:web:example.com%3A3000:user:alice 
+did:web:example.com%3A3000:user:alice
  -> https://example.com:3000/user/alice/did.json
           </pre>
         </section>
@@ -309,21 +309,25 @@ DID:
 
           <ol>
             <li>
+Remove the <code>did:web:</code> prefix.
+            </li>
+            <li>
 Replace ":" with "/" in the method specific identifier to obtain the fully
 qualified domain name and optional path.
             </li>
             <li>
-If the domain contains a port percent decode the colon.
+If the domain contains a port, percent decode the colon.
             </li>
             <li>
 Generate an HTTPS URL to the expected location of the DID document by
 prepending <code>https://</code>.
             </li>
             <li>
-If no path has been specified in the URL, append <code>/.well-known</code>.
+If no path has been specified in the resulting URL, or if the path does not end
+in a <code>/</code> character, append <code>/</code>.
             </li>
             <li>
-Append <code>/did.json</code> to complete the URL.
+Append <code>did.json</code> to complete the URL.
             </li>
             <li>
 Perform an HTTP <code>GET</code> request to the URL using an agent that can
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ https://w3c-ccg.github.io/
 
 ## Editing and building the specification
 
-This specification uses [ReSpec](https://github.com/w3c/respec/) html. Just edit `index.html` and the draft will be rendered correctly.
+This specification uses [ReSpec](https://github.com/w3c/respec/) html. 
+Just edit `index.html` and the draft will be rendered correctly.
 
-When it's time to do a static snapshot, we use the ReSpec UI in the browser to export static HTML.
+When it's time to do a static snapshot, we use the ReSpec UI in the browser to 
+export static HTML.


msporny (commented):
+1, I like the simplification here, and this also feels like it solves (or allows us to dodge) the debates on content negotiation.

index.html
```diff
@@ -261,8 +261,8 @@ <h3>
             </li>
             <li>
 creating the DID document JSON-LD file including a suitable keypair, e.g.,
-using the Koblitz Curve, and storing the <code>did.json</code> file under the
-well-known URL to represent the entire domain, or under the specified path if
+using the Koblitz Curve, and storing the <code>did.json</code> file in the root
+directory to represent the entire domain, or under the specified path if
```
msporny, line 265: Mildly concerned about this clause, but not enough to push back (yet).

index.html
```diff
@@ -293,7 +293,7 @@ <h3>
           </p>
 
           <pre class="example nohighlight" title="Creating the DID with optional path and port">
-did:web:example.com%3A3000:user:alice 
+did:web:example.com%3A3000:user:alice
```
msporny, line 296: Not going to argue in this PR, but I strongly feel that we should shift to this:

```
did:web:example.com/user/alice.json
```

msporny, line 296:
```suggestion
did:web:example.com%3A3000:user:alice.json
```
This PR is suggesting the above, right? If not, might object. Why aren't we just including file extensions?

index.html
```diff
@@ -309,21 +309,25 @@ <h3>
...
 prepending <code>https://</code>.
             </li>
             <li>
-If no path has been specified in the URL, append <code>/.well-known</code>.
+If no path has been specified in the resulting URL, or if the path does not end
+in a <code>/</code> character, append <code>/</code>.
             </li>
             <li>
-Append <code>/did.json</code> to complete the URL.
+Append <code>did.json</code> to complete the URL.
```
msporny, line 330: Oh, hrm... this might be ok... really, we should just be telling people to do the whole file path. Why aren't we doing that?

index.html
dmitrizagidulin: No, the only reason that line is at all present in this PR, is that my code editor (correctly :) ) drops trailing spaces.

index.html
dmitrizagidulin: We're not allowing whole file paths due to the rough consensus in the discussion of PR #47. See, specifically, [Tobias's comment](https://github.com/w3c-ccg/did-method-web/pull/47#issuecomment-988289932):

> At this stage Im a -1 to any proposal that involves requiring the file extension to be leaked into the resulting did web identifier

msporny: I don't think we have rough consensus on that particular point -- this is what I mean, specifically: issue https://github.com/w3c-ccg/did-method-web/issues/52

index.html
```diff
@@ -274,7 +274,7 @@ <h3>
 
           <pre class="example nohighlight" title="Creating the DID">
 did:web:w3c-ccg.github.io
- -> https://w3c-ccg.github.io/.well-known/did.json
+ -> https://w3c-ccg.github.io/did.json
```
awoie, line 277: This is a breaking change to some deployments of did:web. Is this change really needed, or is this just beautification? .well-known is kind of an established pattern for things like this.

awoie (changes requested):
Want to understand why this change is really needed since it would break existing implementations.

index.html
gribneau: +1

I can live with that.

If we're going to simplify, we might as well take it all the way.

I have a roadmap item to support condensed representations at some point next year, but I can cross-reference `.json` via `did:web` with [Also Known As](https://www.w3.org/TR/did-core/#also-known-as) for compatibility.

index.html
gribneau: I agree that `.well-known` is an established pattern, but the consensus here is moving away from established patterns in favor of greater simplification.

Has anyone attempted to add the suffix to the [IANA well known URI registry](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml)?

awoie: Consensus is moving away from established patterns while also breaking backward compatibility? The .well-known is not even visible in DID Docs, VCs etc. Why is this change needed? It doesn't simplify existing did:web implementations. In fact, those people would need to change the file location on their web hosts.

msporny: > This is a breaking change to some deployments of did:web.

did:web is still an experimental specification in the CCG. Breaking changes are expected.

> Why is this change needed? 

To simplify the implementation of the specification and reduce the attack surface of the DID Method. It has become clear that the algorithms for mapping did:web URIs to file path locations are problematic for a number of reasons -- some people don't want to support content negotiation, some people don't want to support .well-known if they can get away from it, some think that not using regular paths was a mistake, etc. Now that there are more people implementing did:web, it's becoming clear that mistakes were made that should be fixed before the rest of the world picks the technology up.

That said, if there are huge deployments of did:web that we don't know about, there is probably a way to inject a compatibility layer into the specification, but the bar for putting those sorts of hacks into the specification should be high.

awoie: > > This is a breaking change to some deployments of did:web.
> 
> did:web is still an experimental specification in the CCG. Breaking changes are expected.

Fair enough.

> 
> > Why is this change needed?
> 
> To simplify the implementation of the specification and reduce the attack surface of the DID Method. It has become clear that the algorithms for mapping did:web URIs to file path locations are problematic for a number of reasons -- some people don't want to support content negotiation, some people don't want to support .well-known if they can get away from it, some think that not using regular paths was a mistake, etc. Now that there are more people implementing did:web, it's becoming clear that mistakes were made that should be fixed before the rest of the world picks the technology up.

Just wondering, why is .well-known increasing the attack surface?

> That said, if there are huge deployments of did:web that we don't know about, there is probably a way to inject a compatibility layer into the specification, but the bar for putting those sorts of hacks into the specification should be high.

I'm not aware of a huge number of did:webs in production.

awoie: or in other words, why would removing .well-known decrease the attack surface?

OR13: @awoie currently the logic can be implemented like this: https://github.com/transmute-industries/verifiable-data/blob/main/packages/did-web/src/index.js

```
const didToUrl = (did) => {
  const regex = new RegExp(
    `did:web:(?<origin>[a-zA-Z0-9/.\\-_]+)(?<path>[a-zA-Z0-9/.:\\-_]*)`
  );
  const match = did.match(regex);
  if (match.groups.path) {
    return `https://${match.groups.origin}${match.groups.path.replace(
      /:/g,
      "/"
    )}/did.json`;
  }
  return `https://${match.groups.origin}/.well-known/did.json`;
};
module.exports = { didToUrl };
```

By removing the other case, it would become:

```
const didToUrl = (did) => {
  const regex = new RegExp(
    `did:web:(?<origin>[a-zA-Z0-9/.\\-_]+)(?<path>[a-zA-Z0-9/.:\\-_]*)`
  );
  const match = did.match(regex);
  if (match.groups.path) {
    return `https://${match.groups.origin}${match.groups.path.replace(
      /:/g,
      "/"
    )}/did.json`;
  }
 throw new Error('Malformed did:web");
};
module.exports = { didToUrl };
```

Its one less case for developers to implement incorrectly, and one less registry with IANA for us to forget to update.

OR13: > I'm not aware of a huge number of did:webs in production.

@tplooker how many did webs you got in production?

kdenhartog: Our domestic COVID pass solution uses did:web as the issuer. I can't state official numbers, but here's a [public article](https://www.stuff.co.nz/national/health/coronavirus/300476004/overseas-new-zealander-left-waiting-23-days-for-vaccine-pass) that can give you an idea of the rough number of passes issued with did:web. 

The issue we'd have with this change is that we're relying on long caching periods in our verifier app to make offline verification possible and this change could require everyone using this app to verify passes in NZ to update their verification app and if they don't it would impact the ability to verify valid passes. Again I can't mention specifics on number of installs, but [Google does publish](https://play.google.com/store/apps/details?id=nz.govt.health.covidpassverifier) on the play store for android there's 100,000+ installs of this app already.

This is just a rough estimate of the size we're operating on with did:web and we're using the .well-known endpoint which makes this change that much more problematic for us. Can we please not intentionally change this unless it's absolutely necessary?

awoie: @OR13 I see, I'm also supportive of simplifying the spec but after hearing @kdenhartog feedback I think we would need good reasons to change it. I'm also interested in @mirceanis feedback since they also have an implementation.

awoie: @OR13 @msporny after hearing @kdenhartog feedback, do you have some thoughts on how backward compatibility could be done after introducing a breaking change like this?

msporny: I'm hearing @kdenhartog say that they have ONE did:web published (one issuer, NZ Govt). Each COVID passport doesn't have a unique did:web identifier associated with it, does it? The ecosystem is also controlled by one implementer (MATTR) and in one country.

If it's just ONE, we could literally hard code it in the specification (worst case), or provide a "Implementations MAY translate known did:web values preceding the specification to ensure backwards compat". There are many ways that we can ensure that we don't break MATTR's implementation.

@kdenhartog What's the URL that you're using for the issuer? (or provide an example of one that looks like it).

kdenhartog: We have one issuer, but in order to resolve the trust gap of knowing that the pass is issued by a trusted issuer we have 100,000+ links to that single issuer which would need to be updated. For us it's not only about the number published but also about the number of back links to that single did:web.

gribneau: I could give you a rewrite rule to maintain compatibility if a change is made to the spec.

kdenhartog: The issue isn't a technical issue. In our code it would probably be quite small to update it. The impact is a deployment problem that's going to leave many non technical people clueless to why they're unable to verify passes that worked yesterday. Furthermore, we don't have an excellent way to notify everyone in the country to say "update your apps because we wanted to change the path of a URL and that's why none of your passes verify anymore".

OR13: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

https://support.cloudflare.com/hc/en-us/articles/200172286-Configuring-URL-forwarding-or-redirects-with-Cloudflare-Page-Rules

Etc... redirects are followed.

msporny: > @msporny takes bets on not all did:web implementations following 302 redirects :)

gribneau: Internal rewrites do not change the apparent URL - the server just returns the specified data as though it were at the target URL. (The DID would appear at both the new and the old location.)

Redirects are also possible, Whether those are allowed or respected depends on the client.

msporny: > The impact is a deployment problem that's going to leave many non technical people clueless to why they're unable to verify passes that worked yesterday.

What would cause them to be unable to verify passes that worked yesterday? Why can't you keep the `did:web` file in the same location as it existed yesterday AND serve it from the new location? Wouldn't this enable full backwards compat for your deployment base?

To be perfectly clear -- I expect everyone here is taking this seriously and we absolutely do not want to break MATTR/NZ's deployment. All we need to do is figure out a path forward that doesn't break the deployment AND fixes some of the issues w/ `did:web`.

gribneau: Note that Apple devices are somewhat famous for creating problems with 30x responses. Verify the best practice if using a redirect - I think it was 307 the last time I checked.

(An internal rewrite is cleaner from a client perspective.)

gribneau: > Why can't you keep the `did:web` file in the same location as it existed yesterday AND serve it from the new location?

If we let webservers be web servers, this conflict is easily resolved.

For example, the rewrite rule below will cause nginx to return the same file for the two common URLs that have been discussed.

```
rewrite ^/.well-known/did.json$ /did.json last;
```

kdenhartog: oh duh! 🤦  Sometimes it's the simplest things that I forgot about. I think the redirect solution could be a simple and low impact (we'd still have to update a few things) change that I think could work. Let me check with others about if they're alright with this. It's going to take some time to build the consensus though so please bare with me while I check with others.

index.html
kdenhartog: I'm not a fan of binding the representation to the subject identifier here either so -1 for me.

index.html
kdenhartog: let''s leave this to a separate issue/PR. I think we're going to need more time to resolve issue #52 faster than we can find consensus for this PR.

index.html
gribneau: You can do a redirect (307) that instructs the client to request the revised URL. This will work or fail depending on the client.

The webserver can also be configured with an internal rewrite that returns the DID file on an alternate URL, so that it appears at both URLs. I would recommend the internal rewrite as it is less risky.

tplooker (changes requested):
I'm away on leave today but just wanted to drop in a quick review to say that I'm -1 to this proposal at this stage for the following reasons

- well known as a convention in part exists to create a dedicated namespace for well known resources to be published that do not risk polluting or closing off the root namespace of a domain. Therefore removing this looses that feature.
- I've not seen an explanation on how hosting the document under a "well known" path could lead to security issues. And to the contrary if the argument is around implementation complexity then I'd say this pr + suggestions documented in #52 creates a far more flexible but also complex syntax supported in did:web which hence is likely to have negative implications around security

OR13:
There are conflicts and objections, I suggest reopening this PR once they are addressed.

PR #60, opened 2022-07-13
clean up several issues and clarify a few sections

mprorock:
Added another example, along side cleaning up a few issues noted in the spec.

Additionally reorganized a few text locations for clarity and to avoid repetition

### DIFF
--- a/index.html
+++ b/index.html
@@ -96,13 +96,10 @@
   <section id="abstract">
     <p>
       DIDs that target a distributed ledger face significant practical
-      challenges in
-      bootstrapping enough meaningful trusted data around identities to
-      incentivize
-      mass adoption. We propose using a new DID method in conjunction with
-      blockchain-based DIDs that allows them to bootstrap trust using a web
-      domain's
-      existing reputation.
+      challenges in bootstrapping enough meaningful trusted data around
+      identities to incentivize mass adoption. We propose using a new DID method
+      in conjunction with blockchain-based DIDs that allows them to bootstrap
+      trust using a web domain's existing reputation.
     </p>
   </section>
   <section id="sotd">
@@ -117,11 +114,9 @@
       </h2>
       <p>
         The Web DID method specification conforms to the requirements specified
-        in the
-        Decentralized Identifiers v1.0 Specification [[DID-CORE]]. For more
-        information about DIDs and DID method specifications, please also see
-        the
-        [[?DID-PRIMER]]
+        in the Decentralized Identifiers v1.0 Specification [[DID-CORE]]. For
+        more information about DIDs and DID method specifications, please also
+        see the [[?DID-PRIMER]]
       </p>
     </section>
     <section id="conformance">
@@ -209,10 +204,8 @@
       </h2>
       <p>
         The target system of the Web DID method is the web host that the domain
-        name
-        described by the DID resolves to when queried through the Domain Name
-        System
-        (DNS).
+        name described by the DID resolves to when queried through the Domain
+        Name System (DNS).
       </p>
 
     </section>
@@ -238,20 +231,15 @@
       <p>
         The method specific identifier is a fully qualified domain name that is
         secured by a TLS/SSL certificate with an optional path to the DID
-        document.
-        The formal rules describing valid domain name syntax are described in
-        [[RFC1035]], [[RFC1123]], and [[RFC2181]].
+        document. The formal rules describing valid domain name syntax are
+        described in [[RFC1035]], [[RFC1123]], and [[RFC2181]].
       </p>
       <p>
         The method specific identifier MUST match the common name used in the
-        SSL/TLS
-        certificate, and it MUST NOT include IP addresses. A port MAY be
-        included and
-        the colon MUST be percent encoded to prevent a conflict with paths.
-        Directories
-        and subdirectories MAY optionally be included, delimited by colons
-        rather than
-        slashes.
+        SSL/TLS certificate, and it MUST NOT include IP addresses. A port MAY be
+        included and the colon MUST be percent encoded to prevent a conflict
+        with paths. Directories and subdirectories MAY optionally be included,
+        delimited by colons rather than slashes.
       </p>
 
       <pre class="nohighlight">
@@ -275,10 +263,9 @@ did:web:example.com%3A3000
       <section>
         <p>
           There is intentionally no HTTP API specified for did:web method
-          operations leaving programmatic
-          registrations and management to be defined by each implementation, or
-          based on their own
-          requirements in their web environment.
+          operations leaving programmatic registrations and management to be
+          defined by each implementation, or based on their own requirements in
+          their web environment.
         </p>
         </p>
         <h3>
@@ -293,24 +280,20 @@ did:web:example.com%3A3000
           </li>
           <li>
             storing the location of a hosting service, the IP address at a DNS
-            lookup
-            service
+            lookup service
           </li>
           <li>
             creating the DID document JSON-LD file including a suitable keypair,
-            e.g.,
-            using the Koblitz Curve, and storing the <code>did.json</code> file
-            under the
-            well-known URL to represent the entire domain, or under the
-            specified path if
-            many DIDs will be resolved in this domain.
+            e.g. using the Koblitz Curve, and storing the <code>did.json</code>
+            file under the well-known URL to represent the entire domain, or
+            under the specified path if many DIDs will be resolved in this
+            domain.
           </li>
         </ol>
 
         <p>
           For example, for the domain name `w3c-ccg.github.io`, the `did.json`
-          will be
-          available under the following URL:
+          will be available under the following URL:
         </p>
 
         <pre class="example nohighlight" title="Creating the DID">
@@ -331,9 +314,8 @@ did:web:w3c-ccg.github.io:user:alice
 
         <p>
           If an optional port is specified on the domain, the port colon
-          splitting the
-          host and the port MUST be percent encoded to prevent collision with
-          the path.
+          splitting the host and the port MUST be percent encoded to prevent
+          collision with the path.
         </p>
 
         <pre class="example nohighlight"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -161,15 +206,18 @@ <h1>
...
+        
         <p>
 The target system of the Web DID method is the web host that the domain name
 described by the DID resolves to when queried through the Domain Name System
 (DNS).
         </p>
+        
+        <p>
+          Unlike [[?DID-DNS]], where the DID is identified as part of a DNS URI 
+          Resource Record, did:web permits both discover of and delivery of a
```
mprorock:
```suggestion
          Resource Record, did:web permits both discovery of and delivery of a 
```

index.html
```diff
@@ -222,6 +270,12 @@ <h2>
 DID method operations
         </h2>
         <section>
+          <p>
+There is no HTTP API specified for did:web method operations process leaving programmatic
```
mprorock:
```suggestion
There is intentionally no HTTP API specified for did:web method operations leaving programmatic
```

index.html
```diff
@@ -161,15 +206,18 @@ <h1>
...
+        
         <p>
 The target system of the Web DID method is the web host that the domain name
 described by the DID resolves to when queried through the Domain Name System
 (DNS).
         </p>
+        
+        <p>
+          Unlike [[?DID-DNS]], where the DID is identified as part of a DNS URI 
+          Resource Record, did:web permits both discovery of and delivery of a
```
OR13: I think we don't need to comment on DID DNS.

OR13 (approved):

index.html
mprorock: agreed - not necessary,  but not harmful (and potentially helpful to folks that are net new to dids [pun intended]) and is a question i have heard asked by multiple folks many times

gribneau: +1

The method is different entirely from `did:dns`.

gribneau (changes requested):
The additional example appears to be `did:key`:

```
  "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```

mprorock:
> The additional example appears to be `did:key`:
> 
> ```
>   "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
> ```

yep - this is pretty common in practice, especially in dev environments - we could change to `did:example` @OR13 thoughts?

OR13:
for did web, I would use `did:web` examples ONLY... no other DID Methods should be referenced imo... unless there is a really good reason.

index.html
```diff
@@ -149,6 +146,54 @@ <h2>
...
+        <pre class="example" title="Example did:web DID document using JWS2020">
+{
+  "@context": [
+    "https://www.w3.org/ns/did/v1",
+    "https://w3id.org/security/suites/jws-2020/v1"
+  ],
+  "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
+  "verificationMethod": [
+    {
+      "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn#z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```
OR13:
```suggestion
      "id": "did:web:example.gov#z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```

OR13 (changes requested):
remove references to did:key

index.html
OR13: this has to match the did.... so maybe don't take my suggestion.

gribneau:
If we're using this additional example to provide clarity to the reader, perhaps explaining that `id` indicates the subject of this document, and that the subject could use a different DID method would have value.

The term `id` is not immediately intuitive as an external reference, but an explanation could highlight interoperability between methods.

OR13:
`id` is the only normative requirement for a did document... and its defined by did core: 

https://www.w3.org/TR/did-core/#did-subject

`id` for a did web, should be a DID, that is of the `did:web:` format.

gribneau:
> `id` is the only normative requirement for a did document... and its defined by did core:
> 
> https://www.w3.org/TR/did-core/#did-subject
>
> `id` for a did web, should be a DID, that is of the `did:web:` format.

Yes, I realize that the DID subject (`id`) is typically the same as the DID example documents, but is that a requirement?

https://www.w3.org/TR/did-core/#also-known-as

In the event that a single subject uses more than one DID, the subject (`id`) in at least one of the associated documents will match the subject, rather than the DID being resolved to that document.

In this scenario, is there any reason that the DID methods used by the subject and multiple identifiers it uses must match?

index.html
```diff
@@ -149,6 +146,54 @@ <h2>
...
 }
         </pre>
+        
+        <pre class="example" title="Example did:web DID document using JWS2020">
+{
+  "@context": [
+    "https://www.w3.org/ns/did/v1",
+    "https://w3id.org/security/suites/jws-2020/v1"
+  ],
+  "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```
mprorock:
```suggestion
  "id": "did:web:example.com#key-0",
```

index.html
mprorock:
```suggestion
      "id": "did:web:example.com#key-0",
```

mprorock: yeah - this is such a common use case wasn't sure how we wanted to handle

index.html
```diff
@@ -149,6 +146,54 @@ <h2>
...
+  "verificationMethod": [
+    {
+      "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn#z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
+      "type": "JsonWebKey2020",
+      "controller": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```
mprorock, line 161:
```suggestion
      "controller": "did:web:example.com#owner",
```

index.html
```diff
@@ -149,6 +146,54 @@ <h2>
...
+      "publicKeyJwk": {
+        "kty": "OKP",
+        "crv": "Ed25519",
+        "x": "0-e2i2_Ua1S5HbTYnVB0lj2Z2ytXu2-tYmDFf8f5NjU"
+      }
+    },
+    {
+      "id": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn#z6LSt8Cke1XG6vthTpdek9xx6a5NKz8gEuYPQHJPhRjfREAC",
```
mprorock:
```suggestion
      "id": "did:web:example.com#key-0",
```

index.html
```diff
@@ -149,6 +146,54 @@ <h2>
...
+      "type": "JsonWebKey2020",
+      "controller": "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
```
mprorock, line 171:
```suggestion
      "controller": "did:web:example.com#owner-2",
```

mprorock:
> remove references to did:key

@OR13 please take a look at the suggestions i just dropped in to see if those make sense to you as an adjustment

index.html
OR13: ^ nope... `id` MUST be a DID, NOT a DID URL!

index.html
OR13: this is ok, you can use a less confusing fragment if you wish.

OR13: weird choice... don't like...

index.html
OR13: not sure why you keep using `#owner`... the spec usually uses `#key-0` when in doubt.

index.html
OR13: IMO, its harmful to connect readers to DID DNS... i'm a -1 to including references to it.

OR13 (changes requested):
Remove reference to DID DNS, fix `id` DID and DID URL issues.

index.html
gribneau: -1 to including references to `did:dns`, +1 to removing.

index.html
```diff
@@ -142,380 +147,543 @@ <h2>
...
+        "kty": "EC",
+        "crv": "P-256",
+        "x": "38M1FDts7Oea7urmseiugGW7tWc3mLpJh6rKe7xINZ8",
+        "y": "nDQW6XZ7b_u2Sy9slofYLlG03sOEoug3I0aAPQ0exs4"
+      }
+    },
+  ],
+  "authentication": [
+    "did:key:z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn#z6MktiSzqF9kqwdU8VkdBKx56EYzXfpgnNPUAGznpicNiWfn",
+    "did:example:123#_TKzHv2jFIyvdTGF1Dsgwngfdg3SH6TpDv0Ta1aOEkw"
```
mprorock:
```suggestion
    "did:web:example.com#key-0",
    "did:web:example.com#key-1"
```

index.html
```diff
@@ -142,380 +147,543 @@ <h2>
...
-The target system of the Web DID method is the web host that the domain name
-described by the DID resolves to when queried through the Domain Name System
-(DNS).
-        </p>
-      </section>
+      <p>
+        Unlike [[?DID-DNS]], where the DID is identified as part of a DNS URI
+        Resource Record, did:web permits both discovery of and delivery of a
+        DID document over HTTPS.
+      </p>
```
mprorock:
```suggestion
```

index.html
OR13: This should be ` "id": "did:web:example.com",`

index.html
OR13: controller MUST be a DID.
controller MUST NOT be a DID URL.

index.html
OR13: same comment, controller must be a DID, not a DID URL

OR13 (changes requested):
Make 

didDocument.id a DID (did:web:example.com)
didDocument.verificationMethod[0].controller a DID  (did:web:example.com)

index.html
mprorock: ok - trying to figure out what we want here - adjust the id to did:web, and keep controller as the did:key?

OR13: controller must be did web.... in general.. mixing controllers for verification methods that don't match the did document subject is legal... but not something that is interoperable.... so don't use examples that do this.

mprorock: excellent - thanks

mprorock:
> Make
> 
> didDocument.id a DID (did:web:example.com) didDocument.verificationMethod[0].controller a DID (did:web:example.com)

believe this is resolved now @OR13

OR13 (approved):

dmitrizagidulin (approved):
:+1:

OR13 (approved):

Issue #52, opened 2021-12-14
Use absolute file paths for did:web DIDs.


msporny:
There is a lot of translation that's done with a did:web URL that feels like it could be simplified. For example, why don't we just require that all DID URL paths are absolute?

```
did:web:did.example/did.json   // site-wide DID
did:web:did.example/alice.json     // specific entity DID
did:web:did.example/some/arbitrary/path/foo.json   // you can place DIDs at arbitrary paths
```

I just checked DID syntax, all the above is valid IIUC... why aren't we just doing that instead? It would allow people to just replace "https://" with "did:web:" to get a valid DID Web URL.

OR13:
@tplooker this would be a breaking change for did web, but IMO something so simple may actually by worth it...

msporny:
> @tplooker this would be a breaking change for did web

I think there is a way for it to not be a breaking change. It'll be ugly/hacky, but we can provide an optional code path where if you detect more than two colons in the DID URL for did:web, you can do the weird replacement thing... but we make it a MAY. Or we could do something like -- dereference, and if you don't get anything back, and there are more than two colons, do the weird replacement thing, and try dereferencing again (and if you get a document back, cache that so future dereferences don't have to go through all that rigmarole).

All that said, I'd really like to avoid doing all that (because it adds complexity and attack surface).

gribneau:
> It would allow people to just replace "https://" with "did:web:" to get a valid DID Web URL.

There is a further positive (imo) implication here. Consider the following cases:

```
did:web:example.com/alice
did:web:example.com/alice/
```

In both cases, the server would be returning an index document if `alice` is a directory. This implicitly enables content negotiation and variable representations, if configured server-side without requiring any complexity in the specification.

> I think there is a way for it to not be a breaking change. It'll be ugly/hacky..

+1 on breaking over hacky.

msporny:
> This implicitly enables content negotiation and variable representations, if configured server-side without requiring any complexity in the specification.

That would certainly be my preference -- stay silent, leave web architecture alone. That said, it seems like @OR13 and @mprorock are both -1s to that?

As an alternative, I suggested the absolute paths thing, where you can be specific about the file extension and therefore, media type... and by doing so, you lock yourself into a media type for that particular DID.

[...]

PR #63, opened 2022-08-22
DID:WEB resolves to DID Resolution Result

### LINKED ISSUE #20: Decide what to do about 'proof' section and other DID Doc Metadata
Signing the DID Document itself (adding an OPTIONAL/recommended `proof` section, for example) would offer integrity protection / move towards the did docs being self-certifying.

Since the `proof` property (as well as other metadata fields such as `created` and `updated`) were removed from the DID Core spec, we need to make a decision on how to handle these in the `did:web` method.

Our options:

1. Put the metadata fields (`proof`, `created` etc) into the DID Document itself.
2. Store not the DID Document in the `.json` file on the web server, but instead embed it in a structure similar to the [DID Resolution Result](https://w3c-ccg.github.io/did-resolution/#did-resolution-result).

Example structure for option 1:

```js
{
	"@context": "https://www.w3.org/ns/did/v1",
	"id": "did:web:example:com",
	"authentication": [{
		"id": "did:web:example:com#keys-1",
		"type": "Ed25519VerificationKey2018",
		"controller": "did:web:example:com",
		"publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
	}],
	"service": [{
		"id": "did:web:example:com#vcs",
		"type": "VerifiableCredentialService",
		"serviceEndpoint": "https://example.com/vc/"
	}],
	"proof": {
		// signature over the whole DID document would go here.
	},
	"created": "2019-03-23T06:35:22Z",
	"updated": "2023-08-10T13:40:06Z"
}
```

Example structure for option 2:

```js
{
	"@context": "https://w3id.org/did-resolution/v1",
	"didDocument": {
		"@context": "https://www.w3.org/ns/did/v1",
		"id": "did:web:example:com",
		"authentication": [{
			"id": "did:web:example:com#keys-1",
			"type": "Ed25519VerificationKey2018",
			"controller": "did:web:example:com",
			"publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
		}],
		"service": [{
			"id": "did:web:example:com#vcs",
			"type": "VerifiableCredentialService",
			"serviceEndpoint": "https://example.com/vc/"
		}],
                "proof": {
                    // signature over the whole DID document would go here.
                }
	},
	"didDocumentMetadata": {
		"created": "2019-03-23T06:35:22Z",
		"updated": "2023-08-10T13:40:06Z"
	}
}
```

vitorpamplona:
Fixes #20: 
1. Changes  the nature of the `did.json` document to become a [DID Resolution Result](https://www.w3.org/TR/did-core/#did-resolution) that contains a [DID Document](https://www.w3.org/TR/did-core/#dfn-did-documents) and not the DID Document itself.
2. Establishes the presence of [DID Document Metadata](https://www.w3.org/TR/did-core/#did-document-metadata) with possibilities of adding cryptographic proofs to verify DID Documents.
3. Allows multiple [representations](https://www.w3.org/TR/did-core/#representations) (JSON, JSONLD, XML, CBOR, etc) of the DID Resolution Document for the same DID:WEB based on [DID Resolution Options](https://www.w3.org/TR/did-core/#did-resolution-options)

### Impact on current implementers

This is a backward incompatible change. 

**Current producers** must wrap the root property of their `did.json` file into a `didDocument` property. 

**DID Resolvers** should check for the presence of a `didDocument` property to indicate compliance of the producer with this new version. If the property is present, an optional `didDocumentMetadata` property can also be available and should be parsed accordingly.  

The spec's `did.json` goes from:  
```json
{
  "@context": "https://www.w3.org/ns/did/v1",
  "id": "did:web:example.com",
  "verificationMethod": [{
     "id": "did:web:example.com#controller",
     "type": "Secp256k1VerificationKey2018",
     "controller": "did:web:example.com",
     "ethereumAddress": "0xb9c5714089478a327f09197987f16f9e5d936e8a"
  }],
  "authentication": [
     "did:web:example.com#controller"
  ]
}
```

To: 

```json
{
  "@context": "https://w3id.org/did-resolution/v1",
  "didDocument": {
    "@context": "https://www.w3.org/ns/did/v1",
    "id": "did:web:example:com",
    "authentication": [{
      "id": "did:web:example:com#keys-1",
      "type": "Ed25519VerificationKey2018",
      "controller": "did:web:example:com",
      "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
    }],
    "service": [{
      "id": "did:web:example:com#vcs",
      "type": "VerifiableCredentialService",
      "serviceEndpoint": "https://example.com/vc/"
    }],
  },
  "didDocumentMetadata": {
    "@context": [
      "https://w3id.org/did-resolution/v1"
      "https://w3id.org/security/suites/ed25519-2020/v1"
    ],
    "created": "2019-03-23T06:35:22Z",
    "updated": "2023-08-10T13:40:06Z",
    "proof": {
      "type": "Ed25519Signature2020",
      "created": "2022-08-21T12:54:33Z",
      "verificationMethod": "did:web:example.com:trustedSigner#WEB",
      "proofPurpose": "assertionMethod",
      "proofValue": "z3e7VcjAVz7RVb7uuH7NQzDuF9AXK3vfH65xnSJSRiLK3vTPUoCDgcE4JC1UfZhbD1wgFsWxQwdWEtrRqhbnqvyYs"
    }
  }
}
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -193,7 +193,7 @@ did:web:example.com%3A3000
       </h2>
       <section>
         <p>
-          There is intentionally no HTTP API specified for did:web method
+          There is intentionally no HTTP API specified for <code>did:web</code> method
           operations leaving programmatic registrations and management to be
           defined by each implementation, or based on their own requirements in
           their web environment.
@@ -225,13 +225,13 @@ did:web:example.com%3A3000
             for the DID Resolution Document
           </li>
           <li>
-            storing the chosen representations of the DID Resolution Document in the <code>did.representation</code> 
-            file under the well-known URL to represent the entire domain, or under the specified path if many DIDs 
+            storing the chosen representations of the DID Resolution Document as <code>did.representation</code> 
+            files under the well-known URL to represent the entire domain, or under the specified path if many DIDs 
             will be resolved in this domain.
           </li>
           <li>
-            setup the HTTP server content negotiation modules to reply with different representations based on 
-            the accept header. As an example: 
+            setting up the HTTP Server Content Negotiation modules to reply with different representations based on 
+            the <code>accept</code> header. As an example: 
             <ol>
               <li><code>accept: application/did+json</code> -> <code>did.json</code></li>
               <li><code>accept: application/did+ld+json</code> -> <code>did.jsonld</code></li>
@@ -253,7 +253,7 @@ accept: application/did+json
 
         <p>
           If the 
-          <a href="https://www.w3.org/TR/did-core/#did-resolution-options">DID Resolution Options's</a> <code>accept</code> 
+          <a href="https://www.w3.org/TR/did-core/#did-resolution-options">DID Resolution Options'</a> <code>accept</code> 
           property is a JSON-LD (<code>application/did+ld+json</code>), it resolves the DID Resolution Document from the following URL:
         </p>
 
@@ -265,7 +265,7 @@ accept: application/did+json
 
         <p>
           If an optional path is specified rather the bare domain, the
-          <code>did.json</code> will be available under the specified path:
+          <code>did</code> file will be available under the specified path:
         </p>
 
         <pre class="example nohighlight" title="Creating the JSON DID with optional path">
@@ -417,7 +417,7 @@ accept: application/did+json
             fully qualified domain name and optional path.
           </li>
           <li>
-            If the domain contains a port percent decode the colon.
+            If the domain contains a port, percent decode the colon.
           </li>
           <li>
             Generate an HTTPS URL to the expected location of the DID Resolution Document
--- a/index.html
+++ b/index.html
@@ -214,8 +214,8 @@ did:web:example.com%3A3000
             lookup service
           </li>
           <li>
-            creating a DID Resolution Document with a <code>didDocument</code>
-            and a <code>didDocumentMetadata</code> property. 
+            creating a DID Resolution Document with a required <code>didDocument</code>
+            and an optional <code>didDocumentMetadata</code> property. 
             The <a href="https://www.w3.org/TR/did-core/#dfn-did-documents">DID Document</a>
             and <a href="https://www.w3.org/TR/did-core/#did-document-metadata">DID Document Metadata</a> 
             data models are defined in the <a href="https://www.w3.org/TR/did-core/">DID-Core</a> specification.
--- a/index.html
+++ b/index.html
@@ -432,7 +432,7 @@ accept: application/did+json
           </li>
           <li>
             Map the <a href="https://www.w3.org/TR/did-core/#did-resolution-options">DID Resolution Option's</a> accept 
-            property to the appropriate request header fields of the HTTP Content Negotiation:
+            property to the appropriate request header fields of the HTTP Content Negotiation
           </li>
           <li>
             Perform an HTTP <code>GET</code> request to the URL with the appropriate headers using an agent
--- a/index.html
+++ b/index.html
@@ -373,7 +373,10 @@ accept: application/did+json
 {
   "@context": "https://w3id.org/did-resolution/v1",
   "didDocument": {
-    "@context": "https://www.w3.org/ns/did/v1",
+    "@context": [ 
+      "https://www.w3.org/ns/did/v1", 
+      "https://w3id.org/security/suites/ed25519-2020/v1"
+    ],
     "id": "did:web:example:com",
     "authentication": [{
       "id": "did:web:example:com#keys-1",
@@ -386,14 +389,6 @@ accept: application/did+json
       "type": "VerifiableCredentialService",
       "serviceEndpoint": "https://example.com/vc/"
     }],
-  },
-  "didDocumentMetadata": {
-    "@context": [
-      "https://w3id.org/did-resolution/v1"
-      "https://w3id.org/security/suites/ed25519-2020/v1"
-    ],
-    "created": "2019-03-23T06:35:22Z",
-    "updated": "2023-08-10T13:40:06Z",
     "proof": {
       "type": "Ed25519Signature2020",
       "created": "2022-08-21T12:54:33Z",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -351,15 +408,22 @@ <h3>
...
           <li>
             If no path has been specified in the URL, append
             <code>/.well-known</code>.
           </li>
           <li>
-            Append <code>/did.json</code> to complete the URL.
+            Map the DID Document Option's accept property to a file extension:
+            <ol>
+              <li><code>application/did+json</code> -> <code>json</code></li>
+              <li><code>application/did+jsonld</code> -> <code>jsonld</code></li>
```
OR13:
```suggestion
              <li><code>application/did+ld+json</code> -> <code>jsonld</code></li>
```

index.html
```diff
@@ -326,13 +269,127 @@ <h3>
...
+  "didDocumentMetadata": {
+    "created": "2019-03-23T06:35:22Z",
+    "updated": "2023-08-10T13:40:06Z",
+    "proof": {
+      "type": "Ed25519Signature2020",
+      "created": "2022-08-21T12:54:33Z",
+      "verificationMethod": "did:web:example.com:trustedSigner#WEB",
+      "proofPurpose": "assertionMethod",
+      "proofValue": "z3e7VcjAVz7RVb7uuH7NQzDuF9AXK3vfH65xnSJSRiLK3vTPUoCDgcE4JC1UfZhbD1wgFsWxQwdWEtrRqhbnqvyYs"
+    }
```
msporny, line 351: If you are trying to express a `proof` for the DID Document, -1 to detaching the `proof` in this way. You can include it in `didDocument`, you just need to add the appropriate context file. If we detach the `proof` in this way, we'll have to have special processing instructions just for did:web resolution which would be special just for did:web. Please move `proof` to the `didDocument` and add the appropriate `@context` definition to make it clean.

If you are trying to digitally sign the response metadata, where `proof` is is fine... but that presumes the DID Resolution context has the `Ed25519Signature2020` declaration OR you need to declare it in the didResolutionMetadata block.

index.html
```diff
@@ -351,15 +408,22 @@ <h3>
...
+            Generate an HTTPS URL to the expected location of the DID Resolution Document
             by prepending <code>https://</code>.
           </li>
           <li>
             If no path has been specified in the URL, append
             <code>/.well-known</code>.
           </li>
           <li>
-            Append <code>/did.json</code> to complete the URL.
+            Map the DID Document Option's accept property to a file extension:
```
msporny: If we are going to do content negotiation, we should follow the HTTP standard: https://www.rfc-editor.org/rfc/rfc7231#section-5.3 or at least ask the HTTP WG what the suggested path is for Conneg these days.

msporny (changes requested):
In general, +1 to this PR -- the response being a DID Resolution result makes a lot of sense.

The only concerning bits are: 1) the way the `proof` is detached from the DID Document (don't do that), and 2) the content negotiation bit.

Regarding concern 1, if you leave the `proof` as a part of the DID Document (and add an appropriate context), that would address that issue. Unless you were really trying to digitally sign the DID Document metadata, in which case, the proof is probably fine where it is?

Regarding concern 2, move the content negotiation changes into a separate PR, as that is a permathread that has been burning for a while: https://github.com/w3c-ccg/did-method-web/issues/52 ...we might just want to ask the HTTP WG what the suggested path is for Conneg these days.

index.html
gribneau: > If we are going to do content negotiation, we should follow the HTTP standard

Strong +1 on following HTTP. DID core is very consistent with HTTP content negotiation.

https://www.w3.org/TR/did-core/#did-resolution-options

https://www.w3.org/TR/did-core/#did-resolution-metadata

peacekeeper:
> -1 to detaching the `proof` in this way. You can include it in `didDocument`

I know that in VCs and Data Integrity, the assumption has always been that the `proof` property is included in the same JSON-LD object that it covers, but I would contrast that with the initial design that we had when we introduced DID metadata (based on this proposal: https://github.com/w3c/did-core/issues/65#issuecomment-597030882). We said at that time that a proof is not data about the subject, but rather metadata about either the document, or the resolution process, and that therefore it belongs into one of the two metadata buckets.

Semantically, having the `proof` inside the DID document would be a bit like having the `proof` inside the `credentialSubject` object, no?

Structurally, having the `proof` outside the DID document would make it easier to cleanly distinguish between proofs that are added by the DID controller or the DID method (DID document metadata), and proofs that are added by a resolver (DID resolution metadata).

From a pure RDF/JSON-LD perspective, I think there shouldn't be a strict necessity to have the `proof` object directly under the object it applies to, or?

peacekeeper (changes requested):
I don't really have a strong opinion whether or not `did:web` should be changed, but if the decision is to support metadata, then this needs to be better aligned with the [DID Resolution specification](https://w3c-ccg.github.io/did-resolution/), e.g.:

- Use of different names for the same thing "DID Resolution Document" vs. "DID Resolution Result".
- The DID Resolution specification already defines an HTTP(S) mapping of the DID Resolution process (e.g. the use of the Accept: header), so there should be no need to re-define these things here.
- This PR seems to introduce a strange new JSON-LD context `https://www.w3.org/ns/didMetadata/v1`, even though `https://w3id.org/did-resolution/v1` already defines terms for metadata.

vitorpamplona:
@peacekeeper Thanks Markus! I fixed them all. 

I think the text could use a lot of improvements outside the context of this PR to be better aligned with the DID Resolution specification. I am just trying to not change everything. :) 

For instance, I think we need to force DNSSec just like we force HTTPS. But that is for another day.

mprorock (changes requested):
Not really interested in breaking changes right now

vitorpamplona:
> Not really interested in breaking changes right now

Hi Mike, any suggestion on how to add the DID Document Metadata with backwards compatibility?

letmaik:
Doesn't this prevent the actual resolver from adding its own metadata?

peacekeeper:
> Doesn't this prevent the actual resolver from adding its own metadata?

There are two "buckets" of metadata. [DID document metadata](https://www.w3.org/TR/did-core/#did-document-metadata) and [DID resolution metadata](https://www.w3.org/TR/did-core/#did-resolution-metadata). Metadata about the resolution process itself should only go into the second bucket.

OR13:
I suggest we close this PR, and perhaps tackle some of the concepts addressed here using additional headers.

OR13:
Closing this PR (it does not seem like it can reach consensus).

I am open to discussing on issues, feel free to file, and link here.

PR #80, opened 2023-08-11
Verify document.id after resolution

### LINKED ISSUE #79: Resolver should check resolved document (ID) matches the expected one
When implementing `did:web` resolution, I noticed the reference implementation checks the ID of the resolved document v.s. the input DID (https://github.com/decentralized-identity/web-did-resolver/blob/master/src/resolver.ts#L40).

This is not mentioned in the `did:web` specification, but should probably be the last resolution step (https://w3c-ccg.github.io/did-method-web/#read-resolve).

reinkrul:
To mitigate (accidentally) misconfigured HTTP servers.

(as implemented by https://github.com/decentralized-identity/web-did-resolver/blob/master/src/resolver.ts#L40)

Fixes https://github.com/w3c-ccg/did-method-web/issues/79

### DIFF
--- a/index.html
+++ b/index.html
@@ -417,7 +417,7 @@ did:web:example.com%3A3000:user:alice
               href="#security-and-privacy-considerations"></a>.
           </li>
           <li>
-            Verify the ID of resolved DID document equals the Web DID being resolved.
+            Verify that the ID of the resolved DID document matches the Web DID being resolved.
           </li>
           <li>
             When performing the DNS resolution during the HTTP <code>GET</code>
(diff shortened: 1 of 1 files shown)


dmitrizagidulin (approved):
:+1: 
I think this additional step makes a lot sense.

index.html
```diff
@@ -416,6 +416,9 @@ <h3>
             enforces the security requirements as described in <a
               href="#security-and-privacy-considerations"></a>.
           </li>
+          <li>
+            Verify the ID of resolved DID document equals the Web DID being resolved.
```
TallTed:
```suggestion
            Verify that the ID of the resolved DID document matches the Web DID being resolved.
```

TallTed (approved):

OR13 (approved):

